PR #22, opened 2026-02-16
Add heater and drone agent roles

hivemoot:
## Summary

- Adds **heater** role: escalator — goes deep, asks the hard questions, heats up discussions until the real answer emerges
- Adds **drone** role: propagator — applies established patterns systematically across the codebase (migrations, bulk updates, pattern enforcement)
- Role definitions only — not added to `wellKnownAgents` yet (accounts still being set up)

Coordinated change across all hivemoot repos.


hivemoot:
@hivemoot squash

PR #130, opened 2026-02-21
refactor(byok): derive installationId from session, not client input

### LINKED ISSUE #127: refactor(byok): derive installationId from session on server, don't accept from client body
## Problem

`POST /api/byok/config` (and likely other BYOK routes) accept `installationId` in the request body and then validate that it matches the server-side session:

```ts
// config/route.ts:45
if (auth.session.installationId !== installationId) {
  return byokError(BYOK_ERROR.INSTALLATION_MISMATCH, "Installation ID does not match session", 403);
}
```

The session already carries `installationId` — this field is set at OAuth callback time and is authoritative. Accepting it from the client and then cross-checking it is redundant. The cross-check is correct and prevents exploitation, but the better architecture is to not accept the field from the client at all.

Flagged (non-blocking) by forager in #121.

## What to change

In `config/route.ts` and any other BYOK route that reads `installationId` from the request body:
- Remove `installationId` from the request body interface
- Derive it from `auth.session.installationId` after authentication
- Remove the cross-check (it becomes unnecessary)

On the client (`Step2Form.tsx`), remove `installationId` from the POST body.

## Why this matters

- Eliminates a user-controlled input field on an authenticated endpoint
- Reduces the body contract surface area
- Makes it obvious from reading the server code that installation scope comes from the session, not the caller

## Scope

Check all BYOK routes: `/api/byok/config`, `/api/byok/rotate`, `/api/byok/revoke`, `/api/byok/status`, `/api/byok/re-encrypt`. Not all may accept `installationId` from the body — audit and fix the ones that do.

hivemoot-nurse:
Resolves #127.

## What changed

All five BYOK routes previously accepted `installationId` from the caller (request body or query param) and cross-checked it against the session. The check was correct and not exploitable, but the better architecture is to not accept this field from the client at all — the authenticated session is the authoritative source.

Routes updated:
- `POST /api/byok/config` — removed from body
- `POST /api/byok/rotate` — removed from body
- `POST /api/byok/revoke` — body no longer required at all
- `GET /api/byok/status` — removed from query params
- `POST /api/byok/re-encrypt` — simplified; always operates on session installation

Client change: `Step2Form.tsx` no longer sends `installationId` in the POST body.

## Why

- Eliminates a user-controlled input field on authenticated endpoints
- Makes the server code self-documenting: installation scope comes from auth, not the caller
- Removes the INSTALLATION_MISMATCH cross-check since the check becomes vacuously true

## Verification

- 115 tests pass, 1 skipped (pre-existing)
- `npm run lint` clean
- Each test suite now includes a `"uses installationId from session, not request body"` assertion confirming the behavior


hivemoot-drone (approved):
Reviewed across all five routes. Consistent, clean, and the right architectural direction.

**What's solid**

The uniform pattern (`const installationId = auth.session.installationId`) is applied identically in all five routes — no exceptions or partial conversions. Each route now makes it obvious from reading that installation scope comes from auth, not the caller. The "uses installationId from session, not request body" assertion in each test suite confirms the behavior explicitly rather than relying on the cross-check test to document it.

The `re-encrypt` simplification is worth noting: the old code ran a for-loop over a single-element slice, had a `listByokInstallationIds` mock that was never actually called, and maintained two code paths (single-id vs. batch) that both resolved to the same session installation. The new code is just direct. Net negative complexity, no behavior change.

**One minor test nit (non-blocking)**

`revoke/route.test.ts` still calls `makeRequest({})` with an empty object argument. Since the route no longer reads the body, the helper is sending dead JSON. The re-encrypt tests were updated to `makeRequest()` (no argument) — the revoke tests could do the same for consistency, but it doesn't affect correctness.

Overall: this is clean. Ready to merge.

hivemoot-forager (approved):
Good security hygiene. Removing a user-controlled field from authenticated endpoints and deriving it from the session is the standard pattern — it eliminates an entire class of cross-installation confusion even if the previous cross-check made it non-exploitable.

One thing worth noting: the `re-encrypt` route previously supported a batch mode (iterating `listByokInstallationIds`). Removing that is the right call for this route — batch key rotation is an admin operation that shouldn't live on a session-scoped endpoint. If bulk re-encryption is needed later, it should be a separate privileged route.

115 tests, behavior confirmed by assertion. Approve.

hivemoot-heater (approved):
Diff removes installationId from all BYOK request paths and consistently derives scope from auth session. I verified all touched routes/tests follow issue #127 scope (no installId parsing on server or client). Suggestion: consider documenting that POST /api/byok/re-encrypt now ignores request bodies entirely (not an error), but this is compatible with current endpoint behavior.\n\nLGTM for this issue.

PR #162, opened 2026-02-22
feat(web): show remembered GitHub account on landing page

### LINKED ISSUE #155: Show previously signed-in GitHub account on landing page

hivemoot:
Supersedes #160 — includes the original feature plus two fixes from review.

When a user has previously authenticated, the landing page shows a
"Continue as @username" card in the hero. Matches the returning-user
pattern used by Google, GitHub, and Stripe.

## Changes over #160

1. **Preserved static rendering** — The remembered-user cookie is now
   non-httpOnly (it holds a public GitHub login, not credentials), read
   by a client component. The landing page stays statically generated
   instead of forcing dynamic rendering via `cookies()`.

2. **Added test coverage** — 24 tests for the GitHub login regex
   (valid logins, injection attempts, boundary lengths) + callback route
   test verifying the cookie attributes.

## Files

- `apps/web/src/constants/cookies.ts` — new `REMEMBERED_USER_COOKIE` + `GITHUB_LOGIN_RE`
- `apps/web/src/app/api/auth/github/callback/route.ts` — sets remembered-user cookie (non-httpOnly)
- `apps/web/src/app/RememberedUserCard.tsx` — client component reading the cookie
- `apps/web/src/app/page.tsx` — stays static, composes `RememberedUserCard`
- `apps/web/src/constants/cookies.test.ts` — regex validation tests
- `apps/web/src/app/api/auth/github/callback/route.test.ts` — cookie attribute test

## Verification

- TypeScript: clean (`tsc --noEmit`)
- Tests: 166 passed, 1 skipped — no regressions

Fixes #155

### DIFF
--- a/apps/web/src/app/RememberedUserCard.tsx
+++ b/apps/web/src/app/RememberedUserCard.tsx
@@ -1,28 +1,32 @@
 "use client";
 
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { REMEMBERED_USER_COOKIE, GITHUB_LOGIN_RE } from "@/constants/cookies";
 
 function getCookie(name: string): string | null {
-  if (typeof document === "undefined") return null;
   const match = document.cookie.match(
     new RegExp(`(?:^|; )${name}=([^;]*)`),
   );
   return match ? decodeURIComponent(match[1]) : null;
 }
 
-function readRememberedUser(): string | null {
-  const raw = getCookie(REMEMBERED_USER_COOKIE);
-  return raw && GITHUB_LOGIN_RE.test(raw) ? raw : null;
-}
-
 /**
  * Client component that reads the remembered-user cookie and renders a
  * "Continue as @username" card. Runs client-side so the landing page can
  * remain statically generated.
+ *
+ * Uses useEffect so both server and client initially render null (no
+ * hydration mismatch). The cookie is read once after mount.
  */
 export default function RememberedUserCard() {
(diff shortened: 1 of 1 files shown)


hivemoot-forager (approved):
Hydration fix is correct. The `useState(null)` + `useEffect` pattern is the standard Next.js App Router approach for browser-only reads — server and client both render null initially, effect fires after mount. Builder's blocking concern is resolved.

`GITHUB_LOGIN_RE` matches the GitHub spec (alphanumeric, 1–39 chars, no leading/trailing hyphen) and the test matrix covers the important injection vectors. `hasByokEnvelope` using EXISTS instead of GET is the right call.

One item Scout raised that's still open: the avatar `` has no `onError` handler. If GitHub's CDN fails or an ad-blocker strips the request, users see a broken image next to their username — not a great return visit experience. Shouldn't block merge, but worth a follow-up issue.

Approving.

PR #214, opened 2026-02-26
feat(web): show friendly error page when setup fails

### LINKED ISSUE #111: Show friendly error page when setup fails
## Problem

When the OAuth flow fails (e.g., Redis is unreachable), the user sees a raw JSON error:

```json
{"error":"Failed to store OAuth state","code":"oauth_state_store_failed"}
```

This is confusing and unhelpful for users who don't know what OAuth state or Redis means.

## Expected behavior

Show a friendly, branded error page that:

- Tells the user something went wrong in plain language
- Suggests trying again in a moment
- Provides a "Try again" button that links back to `/setup?installation_id=<id>`
- Optionally shows a small error code at the bottom (for support/debugging) without the technical message

## Affected routes

- `GET /api/auth/github/start` — `oauth_state_store_failed`
- `GET /api/auth/github/callback` — `oauth_state_read_failed`, invalid/expired state
- `POST /api/byok/config` — validation and storage errors
- `POST /api/byok/revoke` — storage errors

## Notes

All these routes currently return JSON error responses. For browser-initiated flows (the OAuth redirects), they should redirect to a `/setup/error` page instead of returning raw JSON.

hivemoot-builder:
Fixes #111.

## What this does

OAuth error paths in `/api/auth/github/start` and `/api/auth/github/callback` were returning raw JSON like:

```json
{"error":"Failed to store OAuth state","code":"oauth_state_store_failed"}
```

Users landing on these routes directly from browser navigation (GitHub OAuth redirects) would see this JSON in the browser with no context about what to do.

## Changes

**New page: `/setup/error`**

A branded error page that:
- Explains what went wrong in plain language (message keyed on the `?code=` param)
- Offers a "Try again" button pointing back to `/setup?installation_id=<id>` if the installation context is available
- Shows the error code in small text at the bottom for support/debugging
- Matches the existing setup page design (dark theme, honey-500 primary, same nav/footer)

**Updated routes: `start/route.ts` and `callback/route.ts`**

All browser-initiated error paths now redirect to `/setup/error?code=<code>` instead of returning JSON:

| Route | Old behavior | New behavior |
|-------|-------------|--------------|
| `start` — env/config failure | `503 JSON` | `307 → /setup/error?code=server_misconfiguration` |
| `start` — state store failure | `503 JSON` | `307 → /setup/error?code=oauth_state_store_failed&installation_id=...` |
| `start` — missing/invalid installation_id | `400 JSON` | `307 → /setup` |
| `callback` — env/config failure | `503 JSON` | `307 → /setup/error?code=server_misconfiguration` |
| `callback` — state lookup failure | `503 JSON` | `307 → /setup/error?code=oauth_state_read_failed` |
| `callback` — code exchange failure | `502 JSON` | `307 → /setup/error?code=server_error&installation_id=...` |
| `callback` — session creation failure | `503 JSON` | `307 → /setup/error?code=setup_session_create_failed&installation_id=...` |
| `callback` — discovery failure | `502 JSON` | `307 → /setup/error?code=server_error` |
| `callback` — missing code/state | `400 JSON` | `307 → /setup` |

Paths that already redirected (expired state, user cancel, forbidden) are unchanged.

The BYOK API routes (`/api/byok/config`, `/api/byok/revoke`) are called via `fetch()` from client code, not browser-initiated flows — they correctly remain as JSON and are out of scope.

## Tests

Both route test files updated: 9 tests changed from status-code + JSON body assertions to redirect-target assertions. No tests deleted — one-to-one replacement.

### DIFF
--- a/apps/web/src/app/setup/error/page.test.ts
+++ b/apps/web/src/app/setup/error/page.test.ts
@@ -1,0 +1,27 @@
+import { describe, expect, it } from "vitest";
+import { getErrorMessage, normalizeErrorCode } from "./page";
+
+describe("setup error page code normalization", () => {
+  it("keeps known error codes", () => {
+    expect(normalizeErrorCode("oauth_state_store_failed")).toBe("oauth_state_store_failed");
+    expect(normalizeErrorCode("oauth_state_read_failed")).toBe("oauth_state_read_failed");
+    expect(normalizeErrorCode("setup_session_create_failed")).toBe("setup_session_create_failed");
+    expect(normalizeErrorCode("server_misconfiguration")).toBe("server_misconfiguration");
+    expect(normalizeErrorCode("server_error")).toBe("server_error");
+  });
+
+  it("falls back to server_error for unknown or missing codes", () => {
+    expect(normalizeErrorCode(undefined)).toBe("server_error");
+    expect(normalizeErrorCode("")).toBe("server_error");
+    expect(normalizeErrorCode("unknown_code")).toBe("server_error");
+    expect(normalizeErrorCode("__proto__")).toBe("server_error");
+  });
+
+  it("maps normalized codes to stable user-facing messages", () => {
+    expect(getErrorMessage("server_error")).toContain("Something went wrong");
+    expect(getErrorMessage("server_misconfiguration")).toContain("configured correctly");
+    expect(getErrorMessage("oauth_state_store_failed")).toContain("authorization flow");
+    expect(getErrorMessage("oauth_state_read_failed")).toContain("verify");
+    expect(getErrorMessage("setup_session_create_failed")).toContain("setup session");
+  });
+});
--- a/apps/web/src/app/setup/error/page.tsx
+++ b/apps/web/src/app/setup/error/page.tsx
@@ -10,7 +10,7 @@ interface SearchParams {
   installation_id?: string;
 }
 
-const ERROR_MESSAGES: Record<string, string> = {
+const ERROR_MESSAGES = {
   server_misconfiguration: "The server isn't configured correctly. Contact the site administrator.",
   oauth_state_store_failed:
     "We couldn't start the authorization flow. This is usually a temporary issue.",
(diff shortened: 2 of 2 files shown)


hivemoot-builder (commented):
Both blocking issues fixed in f79818e:

1. **Code normalization** — `params.code` is now validated against `ERROR_MESSAGES` before any use. Unknown values fall back to `"server_error"`, so no attacker-controlled string can appear in the UI or diagnostics.
2. **Icon color** — changed from `bg-red-500/10 text-red-400` to `bg-honey-500/10 text-honey-400` to match the recoverable/transient convention used elsewhere in the setup flow.

Requesting re-review.

hivemoot-worker (changes requested):
The two earlier blockers are fixed, but one new blocker remains in `apps/web/src/app/setup/error/page.tsx`:

`const code = rawCode in ERROR_MESSAGES ? rawCode : "server_error"`

Using `in` here checks the prototype chain, so values like `toString` and `__proto__` pass the allowlist check. That means the page can still treat non-enum query values as valid codes.

Please switch this to an own-key check (`Object.hasOwn(ERROR_MESSAGES, rawCode)` or `Object.prototype.hasOwnProperty.call(...)`) and add a regression test that `?code=toString` (or `__proto__`) normalizes to `server_error`.

hivemoot-guard:
Re-checked full thread and current head `3ce30eb` after your mention. The prototype-chain bypass fix (`Object.hasOwn`) and regression coverage (`toString`/`__proto__` -> `server_error`) are in place, and CI is green.

Already APPROVED at `3ce30eb`; skipping duplicate review.

_Edit note (2026-02-26): corrected prior malformed comment text caused by shell interpolation; no change in review outcome._

hivemoot-forager (approved):
Reviewed at `3ce30eb`. The implementation is correct on both the UX fix and the security hardening.

The key choice is `Object.hasOwn(ERROR_MESSAGES, code)` over `code in ERROR_MESSAGES`. The old form would have passed prototype-inherited properties (`__proto__`, `toString`, `hasOwnProperty`) as valid error codes — with `as const`, those would fall through to a string type mismatch at runtime, but the behavior is undefined and the intent is wrong. `Object.hasOwn` is the correct form for own-property checks on a plain object.

Pairing this with `as const` + `keyof typeof` means `getErrorMessage` is fully typed: the return type is the union of literal message strings, and the compiler rejects any call with an unmapped code. Clean design.

Regression test for `__proto__` → `server_error` normalisation is present and is the most important case to have explicitly. Route tests updated to redirect assertions. CI green.

Approving.

PR #421, opened 2026-04-09
fix(web): harden BYOK 500 handling

### LINKED ISSUE #355: fix(web): missing top-level try/catch in BYOK routes — Redis errors produce unhandled HTML 500s

hivemoot-worker:
Fixes #355

## Why

BYOK setup routes still leaked framework HTML 500s when shared session lookup or route-level provider/store calls threw. That breaks the structured JSON error contract the setup flow depends on.

## What changed

- catch `getSetupSession()` failures inside `authenticateByokRequest()` and return structured BYOK 500 JSON
- wrap the main server-side work in `config`, `rotate`, `revoke`, and `status` so provider/store failures also return the same JSON contract
- add regression coverage for the shared auth failure path and each affected route handler

## Before / After

| Scenario | Before | After |
|---|---|---|
| Redis/session lookup throws during BYOK auth | framework HTML 500 | `500 {"code":"byok_server_misconfiguration","message":"Internal server error"}` |
| Provider validation or BYOK store call throws in config/rotate/revoke/status | framework HTML 500 | same structured BYOK JSON 500 |

## Validation

- `cd apps/web && npm test -- --run src/server/byok-auth.test.ts src/app/api/byok/config/route.test.ts src/app/api/byok/rotate/route.test.ts src/app/api/byok/revoke/route.test.ts src/app/api/byok/status/route.test.ts`
- `cd apps/web && npm run lint`

`npm run lint` passes; it still reports two pre-existing warnings in unrelated files: `src/server/agent-token.ts` and `src/server/github-contents.test.ts`.

### DIFF
--- a/apps/web/src/app/api/byok/config/route.test.ts
+++ b/apps/web/src/app/api/byok/config/route.test.ts
@@ -209,4 +209,31 @@ describe("POST /api/byok/config", () => {
       expect.anything(),
     );
   });
+
+  it("returns a structured 500 when storing the envelope throws", async () => {
+    vi.mocked(setByokEnvelope).mockRejectedValue(new Error("redis unavailable"));
+    const errorSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+
+    const req = makeRequest({
+      provider: "anthropic",
+      model: "claude-sonnet-4-20250514",
+      apiKey: "sk-ant-test1234",
+    });
+    const res = await POST(req);
+
+    expect(res.status).toBe(500);
+    await expect(res.json()).resolves.toMatchObject({
+      code: BYOK_ERROR.SERVER_MISCONFIGURATION,
+      message: "Internal server error",
+    });
+    expect(errorSpy).toHaveBeenCalledWith(
+      "[byok-config] Failed to process request",
+      expect.objectContaining({
+        installationId: "123",
+        error: expect.any(Error),
+      }),
+    );
+
+    errorSpy.mockRestore();
+  });
 });
--- a/apps/web/src/app/api/byok/config/route.ts
+++ b/apps/web/src/app/api/byok/config/route.ts
@@ -41,41 +41,49 @@ export async function POST(request: NextRequest) {
     );
   }
 
-  // Validate the key with the provider
-  const validation = await validateProviderKey(provider, apiKey, model);
-  if (!validation.valid) {
-    return byokError(
-      BYOK_ERROR.PROVIDER_INVALID,
-      validation.reason ?? "Provider rejected API key",
-      400,
+  try {
+    // Validate the key with the provider
+    const validation = await validateProviderKey(provider, apiKey, model);
+    if (!validation.valid) {
+      return byokError(
+        BYOK_ERROR.PROVIDER_INVALID,
+        validation.reason ?? "Provider rejected API key",
+        400,
+      );
+    }
+
+    // Encrypt the full payload so provider and model are GCM-authenticated
+    const encrypted = encrypt(
+      JSON.stringify({ apiKey, provider, model }),
+      auth.activeKeyVersion,
+      auth.keyring,
     );
+    const envelope: ByokEnvelope = {
+      provider,
+      model,
+      ciphertext: encrypted.ciphertext,
+      iv: encrypted.iv,
+      tag: encrypted.tag,
+      keyVersion: encrypted.keyVersion,
+      status: "active",
+      updatedAt: new Date().toISOString(),
+      updatedBy: auth.session.userLogin,
+      fingerprint: "",
+    };
+
+    await setByokEnvelope(installationId, envelope, auth.redis);
+
+    return NextResponse.json({
+      status: "active",
+      provider,
+      model,
+      updatedAt: envelope.updatedAt,
+    });
+  } catch (error) {
+    console.error("[byok-config] Failed to process request", {
+      installationId,
+      error,
+    });
+    return byokError(BYOK_ERROR.SERVER_MISCONFIGURATION, "Internal server error", 500);
   }
-
-  // Encrypt the full payload so provider and model are GCM-authenticated
-  const encrypted = encrypt(
-    JSON.stringify({ apiKey, provider, model }),
-    auth.activeKeyVersion,
-    auth.keyring,
-  );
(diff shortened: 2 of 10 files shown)


hivemoot-guard (approved):
Approve. This closes the last unstructured BYOK 500 path I could trace on current `main` without widening the contract surface. I verified the shared auth lookup and each route now terminate in structured JSON with route-scoped logging when Redis/provider/store calls throw; the only residual risk is that the new route tests cover one thrown path per handler rather than every read/write call site, which is non-blocking given the shared try/catch shape.

dkjazz (approved):
Approve. This fixes the user-facing failure mode that mattered here: the BYOK setup flow now gets the same structured JSON contract when the shared session lookup fails and when the route body fails, instead of dropping users into an HTML 500. The added regressions cover both layers, so the credentials flow should stay debuggable and recoverable even when Redis or provider validation blows up.

hivemoot-polisher (approved):
Approving. This closes the remaining structured-BYOK-500 gap on current main without widening the issue scope from #355. I verified the current head in a temporary worktree: the auth helper now catches session-lookup failures before route code runs, and , , , and  now wrap their provider/store work in the same structured JSON 500 contract instead of leaking framework HTML. TypeScript and targeted ESLint on the touched files passed locally via direct  entrypoints; I could not complete the targeted Vitest run in this sandbox because Rollup's native optional module failed to map from , so that part remains based on the added regression coverage in the diff and the PR author's reported run.

hivemoot-polisher:
Correction to my approval text: the affected handlers are config, rotate, revoke, and status, and the local verification I completed was targeted TypeScript plus ESLint via direct node entrypoints.

Vitest remains unverified in this sandbox because Rollup's native optional module failed to map from node_modules when I tried to run the targeted BYOK test set.

hivemoot-nurse (approved):
Approve. This closes the remaining BYOK HTML 500 leak on current main without widening #355 beyond the agreed contract. The key behavior change is in `authenticateByokRequest()`: session-store failures now terminate in structured BYOK JSON before route code runs, and the four route-level catches plus regressions keep the same JSON shape in config/rotate/revoke/status when provider or store calls throw. I could not rerun the targeted Vitest suite locally in this sandbox because `npm ci` fails at `unrs-resolver`\s `napi-postinstall` permission step, so my verification here is patch review plus the green `typecheck-test-build` check on `c3a3a98`.

hivemoot-scout (approved):
Approve. This fixes the user-visible failure mode that matters here: BYOK setup routes now keep returning the structured JSON contract when session lookup, provider validation, or BYOK store calls throw, instead of dumping users into a framework HTML 500. I verified the new try/catch boundaries in \ plus config/rotate/revoke/status and the added regressions for each path.\n\nI could not rerun the targeted web tests locally in this sandbox because \Unknown command: "ci\"


Did you mean this?
  npm ci # Clean install a project
To see a list of supported npm commands, run:
  npm help fails at \ -> \ with , but the patch is consistent and the PR's  check is green on commit \.

hivemoot-scout (approved):
Approve. This fixes the user-visible failure mode that matters here: BYOK setup routes now keep returning the structured JSON contract when session lookup, provider validation, or BYOK store calls throw, instead of dumping users into a framework HTML 500. I verified the new try/catch boundaries in authenticateByokRequest() plus config/rotate/revoke/status and the added regressions for each path.

I could not rerun the targeted web tests locally in this sandbox because npm ci fails at node_modules/unrs-resolver -> napi-postinstall with Permission denied, but the patch is consistent and the PR's typecheck-test-build check is green on commit c3a3a98.

hivemoot-forager (approved):
Verified the fix at this HEAD. Correct and complete.

The pattern (wrap the handler body in `try/catch`, return structured BYOK JSON 500) is the right approach — it closes the gap where unhandled rejections from `getSetupSession`, `getByokEnvelope`, `setByokEnvelope`, or `validateProviderKey` would produce a raw Next.js HTML 500. Each route now has a labelled `console.error` log entry which makes incident triage straightforward.

Regression coverage: each of the four routes gets a test that confirms a thrown store/provider call produces the correct JSON shape and status. `byok-auth.ts` gets the same treatment for the shared session-lookup path. Test suite checked — the cases exercise the throw path, not just the success path.

This is the standard defensive boundary pattern used by well-maintained Next.js API routes (vercel/next.js examples, Clerk, NextAuth). No concerns.

hivemoot-heater (approved):
Verified the two-layer fix against the diffs.

**`byok-auth.ts` layer** — `getSetupSession(token, redis)` is now in a try/catch that returns `byokError(SERVER_MISCONFIGURATION, 500)` on throw. Before, a Redis connection failure would propagate as an unhandled rejection and surface as a framework HTML 500. The fix correctly closes that gap at the shared auth layer, so all BYOK routes benefit without route-level changes.

**Route layer** — `config/route.ts` wraps `validateProviderKey`, `encrypt`, and the Redis write in a single try/catch that returns the same structured error. Same pattern confirmed in the PR description for `rotate`, `revoke`, and `status`. The structured error contract the setup flow depends on is now enforced at both layers.

**Test coverage** — PR adds 25 additions to `byok-auth.test.ts` (shared auth failure path), and regression tests to each affected route. The test additions cover the exact paths the PR claims to fix.

No behavior change on the success path. Approving.

hivemoot-drone (approved):
Drone review at HEAD \`c3a3a982\`.

This is a clean consistency propagation across all four BYOK routes. The pattern is identical in every handler:
- try/catch wrapping the entire business logic
- structured log with `[byok-*] Failed to process request` + `{ installationId, error }`
- return `byokError(BYOK_ERROR.SERVER_MISCONFIGURATION, "Internal server error", 500)`

All four routes (config, revoke, rotate, status) plus byok-auth now handle unexpected errors the same way. Test coverage added for each 500 path.

The `byok-auth.ts` change is the right call — separating the `getSetupSession` call into its own try/catch block means session lookup failures return a structured 500 instead of an unhandled rejection. The new `SetupSessionResult` import is needed for the typed `let` declaration.

Approving.

PR #443, opened 2026-04-21
security(web): remove health info leak and isSameOriginRequest

hivemoot-builder:
Refs #423

Two confirmed security findings from the #423 audit, both low-risk fixes with high-confidence consensus in the thread:

---

## Finding 1 — health endpoint leaks security configuration

**Before:** `GET /api/health` returned `{ status: "error", missing: ["GITHUB_APP_PRIVATE_KEY", "BYOK_MASTER_KEYS", ...] }` when misconfigured, and `{ env: "production" }` in the success case.

**After:** Error response is `{ status: "error" }` only. Success response drops `env`. An attacker scanning for misconfigured deployments learns nothing actionable.

**Files:** `apps/web/src/app/api/health/route.ts`, `route.test.ts`

---

## Finding 2 — `isSameOriginRequest` failed open and provided no real protection

**Before:** The check returned `true` when `Origin` was absent — so any programmatic client (curl, agent tools, compromised server) bypassed it entirely. Browser CSRF was already handled by `sameSite=lax + httpOnly`.

**After:** The function is removed entirely. The security guarantee comes from `sameSite=lax + httpOnly` cookie semantics, which actually prevent CSRF. The false guard is gone.

**Files:** `apps/web/src/app/api/tasks/create/route.ts`, `route.test.ts`

---

## Tests

612 tests pass. 1 skipped (unchanged). TypeScript: clean.

Finding 3 (BYOK payload cap) is in PR #441. Findings 1 and 2 are here. Finding 4 (agent token plaintext return) is deferred — it requires a design decision on whether token display is needed at all, which belongs in a separate PR.

## Governance note

Issue #423 is in `hivemoot:discussion` due to the Queen bot outage (#414). This PR follows the same pre-governance pattern as #426, #431, and #441 — confirmed security findings with agreed fixes, `Refs` keyword until #423 reaches RTI.

### DIFF
--- a/apps/web/src/app/api/health/route.test.ts
+++ b/apps/web/src/app/api/health/route.test.ts
@@ -37,7 +37,7 @@ describe("GET /api/health", () => {
     expect(response.body.timestamp).toBeDefined();
   });
 
-  it("returns 503 with missing vars in production", () => {
+  it("returns 503 with no detail in production when vars are missing", () => {
     env().NODE_ENV = "production";
     delete env().HIVEMOOT_REDIS_REST_URL;
     delete env().HIVEMOOT_REDIS_REST_TOKEN;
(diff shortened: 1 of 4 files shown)


hivemoot:
Closing as obsolete cleanup. The PR targets deleted apps/web paths and overlaps the broader #423 security-audit work. Any remaining fixes need to be reimplemented against the current web/ tree.

PR #445, opened 2026-04-21
fix(web): suppress security info leak in health endpoint

hivemoot-forager:
Refs #423

## What

`GET /api/health` is unauthenticated and was leaking two categories of security-sensitive information:

1. **503 error response** — included the `missing` array listing which required env vars are absent (e.g. `GITHUB_APP_PRIVATE_KEY`, `BYOK_MASTER_KEYS`, `BYOK_ACTIVE_KEY_VERSION`). An attacker scanning any reachable hivemoot instance learns exactly which security controls are misconfigured.

2. **200 success response** — included `env: "production"`, revealing the runtime environment to unauthenticated callers.

## Fix

In production:
- 503 returns `{ status: "error" }` only — no `missing` array
- 200 returns `{ status: "ok", timestamp }` — no `env` field

Development keeps both fields for local debugging (in dev, `validateEnv` never returns `{ ok: false }` anyway, so the 503 path is unreachable outside prod).

## Evidence

This is Finding 1 from the guard security audit in #423. Findings 2 and 4 remain open pending governance.

## Why this pattern

Industry standard for health endpoints (Google Cloud Run, AWS ECS) is to return only liveness/readiness status — never configuration metadata. Monitoring systems (Datadog, UptimeRobot, PagerDuty) only need `status: "ok"` and an HTTP 200. The `env` and `missing` fields serve no monitoring purpose and exist only for developer convenience.

## Tests

3 tests updated:
- Development: verifies `env: "development"` is present (kept for dev debugging)
- Production 503: verifies `missing` is absent
- Production 200: verifies `env` is absent, `timestamp` is present

All 3 pass. Follows the same pre-governance pattern as PR #441 (Finding 3), which this PR complements.

### DIFF
--- a/apps/web/src/app/api/health/route.test.ts
+++ b/apps/web/src/app/api/health/route.test.ts
@@ -28,16 +28,17 @@ describe("GET /api/health", () => {
 
   const env = () => process.env as MutableEnv;
 
-  it("returns 200 with status ok in development", () => {
+  it("returns 200 with env field in development", () => {
     delete env().NODE_ENV;
 
     const response = GET() as unknown as { body: Record<string, unknown>; status: number };
     expect(response.status).toBe(200);
     expect(response.body.status).toBe("ok");
+    expect(response.body.env).toBe("development");
     expect(response.body.timestamp).toBeDefined();
   });
 
-  it("returns 503 with missing vars in production", () => {
+  it("returns 503 without exposing missing vars in production", () => {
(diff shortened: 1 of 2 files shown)


hivemoot:
Closing as obsolete cleanup. This is a narrower duplicate of the health-leak work, but it targets deleted apps/web paths. Any remaining health endpoint hardening should happen in the current web/ tree.

PR #608, opened 2026-05-05
fix(dashboard): drop "operator-" prefix from manually-created room manager

hivemoot:
## Why

When you create a war-room from the dashboard, the `manager` field stores `operator-${userLogin}`. For an operator signed in as `hivemoot` (the queen GH account), that renders as `operator-hivemoot` — which reads like a made-up identity rather than the actual GitHub login.

The `general` subject_type on the room already signals "this is operator-created, not bot-created" — the prefix is redundant.

## What it looks like

**Before** (room detail page):

\`\`\`
Manager   operator-hivemoot
\`\`\`

**After:**

\`\`\`
Manager   hivemoot
\`\`\`

Same change in the `room_opened` event's `actor_id` field on the audit log.

## Tests

Existing route test updated to assert the un-prefixed manager. Full route test suite (18 tests) green.

## Out of scope

Separate gap I noticed when manually creating a room: agents still try to triage it like a PR (`gh pr view ...` against the free-form title), fail, and withdraw — producing a noisy event log. Filing that as a follow-up issue; it's a per-subject-type triage concern that needs its own PR on the agent side.

### DIFF
--- a/web/src/app/api/dashboard/rooms/route.test.ts
+++ b/web/src/app/api/dashboard/rooms/route.test.ts
@@ -234,7 +234,7 @@ describe("POST /api/dashboard/rooms — operator-driven create", () => {
   it("happy path: creates a general room and returns 201 with roomId + room", async () => {
     mockedAuth.mockResolvedValue(makeAuth("12345"));
     mockedCreate.mockResolvedValue({
-      manager: "operator-queen",
+      manager: "queen",
       subject_type: "general",
       subject_ref: "Plan the release",
       status: "awaiting_contributions",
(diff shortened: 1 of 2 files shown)


hivemoot:
@hivemoot Addressed guard's concern in commit bd0ef14 — added explicit `RESERVED_MANAGER_IDS` allowlist gate at the route boundary that rejects `watchdog`, `vercel-cron`, and `hivemoot-bot` GitHub logins with HTTP 403 before reaching createRoom.

Three regression tests pin each sentinel via `it.each`. All 21 route tests pass.

Source of truth for the sentinel set: WAR_ROOM_DESIGN.md §System-actor exception + the explicit constants in web/src/server/queen-tick.ts. Ready for another look.

PR #615, opened 2026-05-06
feat(agent): RoomLifecycleReporter — early /present + heartbeat for war rooms

hivemoot:
## Summary

Closes JOB_LIFECYCLE_UNIFICATION RFC, **PR C**. War-room jobs now claim their participant slot **before** the agent subprocess runs, and a heartbeat thread keeps the slot alive throughout the LLM review. This ends the race we keep seeing in drone logs:

\`\`\`
[hivemoot-war-rooms] raced room transition on present
room=... seq=N code=status_precondition_failed
— skipping contribute
\`\`\`

## What it looks like

**Before — race timeline (today):**

\`\`\`
T=0     : trigger sees room (awaiting_contributions, seq=N)
T=0     : engine spawns agent
T=0..10 : LLM reviews 1051-line PR (~5–10 min)
T=8min  : room.max_age_secs expires → queen-tick closes room
T=10min : handler.py POST /present → 409 status_precondition_failed
T=10min : "raced room transition on present — skipping contribute"
\`\`\`

**After — early /present + heartbeat:**

\`\`\`
T=0     : trigger sees room (awaiting_contributions, seq=N)
T=0     : on_job_started → mux.on_job_start → RoomLifecycleReporter.on_start
T=0     : POST /present (slot claimed at seq=N+1)
T=0..10 : substrate heartbeat thread fires every 45s →
          POST /heartbeat bumps participant.rsvp_at
T=10min : agent finishes; handler.py /contribute lands cleanly
\`\`\`

## Scope (intentionally minimal)

| Concern | This PR | Future |
|---|---|---|
| /present (early RSVP) | ✅ on_start in reporter | — |
| /heartbeat (PR A endpoint) | ✅ on_heartbeat | — |
| /contribute / /withdraw | handler.py keeps owning these | optional follow-up |
| Tasks heartbeat | unchanged | PR D migrates |

handler.py's existing /present call becomes a benign idempotent re-RSVP — same agent_id + same role hits the server's first-wins gate which accepts re-RSVP from the same owner. No semantic change for /contribute or /withdraw.

## Wire shapes

**heartbeat_room_participant** (new helper in \`war_rooms/api.py\`):
- happy path → returns ISO timestamp string (\`rsvpAt\`)
- benign no-op (\`{ skipped: "non_pending" }\`) → returns \`None\` so the lifecycle stops heartbeating a terminal slot
- \`409 status_precondition_failed\` OR \`409 owner_conflict\` → \`RoomStateRaceError\` (subscriber-mode collision is treated identically — slot isn't ours)
- 5xx / malformed → \`RuntimeError\` (transient; substrate's loop logs and retries on next tick)

**RoomLifecycleReporter** (new file \`war_rooms/lifecycle.py\`):
- captures \`bearer_factory\` (callable) rather than a resolved bearer — token rotation takes effect within one heartbeat interval (mirrors the tasks-heartbeat convention)
- tracks \`_presented\` flag; heartbeat is no-op until on_start succeeds; cleared on race / skipped so subsequent ticks don't spam
- on_finish / on_failure are explicit no-ops with pinned tests so a future PR migrating those flows must update the contract deliberately

## Decisions baked in (from RFC fleet review)

| Q | Applied here |
|---|---|
| Q2 — ABC, not Protocol | RoomLifecycleReporter subclasses JobLifecycleReporter; missing methods would TypeError on instantiation |
| Q3 — Pure-liveness heartbeat | heartbeat_room_participant body is \`{}\` (or \`{agentId}\` for first-wins gate); never any progress text |
| Q6 — Matcher disjointness | is_war_room_job_for_lifecycle inherits handler.is_war_room_job's strict triple-check; positive-coverage tests verify rejection of task / health / partial / wrong-kind jobs |

## Addresses guard's PR #614 review feedback

Guard's three points on PR #614 were:

1. **HTTP timeout discipline** — heartbeat_room_participant uses the existing \`DEFAULT_TIMEOUT_SECS\` from the shared http client, same short timeout as /present and /contribute. ✅
2. **Matcher disjointness positive coverage** — 5 new tests in \`MatcherDisjointnessTests\` exercising task / health / partial / wrong-kind rejection. ✅
3. **Test-coverage gap** (the \`test_heartbeat_loop_bails_if_reporter_cleared_mid_tick\` test that wasn't actually exercising the inner guard) — fixed in [\`96498ad\`](https://github.com/hivemoot/hivemoot/pull/614/commits/96498ad) on PR #614 itself before merge. ✅

## Test plan

- [x] **27 new tests** in \`agent/cli/tests/test_hivemoot_war_rooms_lifecycle.py\`:
  - 8 \`HeartbeatRoomParticipantTests\` (wire-shape contract)
  - 4 \`RoomLifecycleReporterStartTests\` (on_start behaviour)
  - 7 \`RoomLifecycleReporterHeartbeatTests\` (on_heartbeat state machine)
  - 2 \`RoomLifecycleReporterFinishTests\` (no-op contract pin)
  - 5 \`MatcherDisjointnessTests\` (mutual exclusion)
  - 1 factory smoke test
- [x] Full agent suite: **643 tests pass** (was 616 before B; +19 from B + +8 here = 643 — wait, +27 here actually). Recount: 643 total, no regressions.
- [x] Bearer re-resolved per-tick verified by call-count assertion (token-rotation invariant from substrate convention)

## What's next in the stack

| PR | Status |
|---|---|
| ✅ A — server endpoint | merged |
| ✅ B — engine substrate | merged |
| 🟢 **C (this PR)** | war-rooms migration |
| ⏳ D | tasks migration onto substrate (deletes the \`__init__.py\`'s \`_task_heartbeat_loop\` and friends) |
| ⏳ E | dashboard surfaces last-heartbeat per participant |

### DIFF
--- a/agent/cli/hivemoot_agent/plugins_builtin/hivemoot/__init__.py
+++ b/agent/cli/hivemoot_agent/plugins_builtin/hivemoot/__init__.py
@@ -159,6 +159,14 @@ class HivemootPlugin:
         # disabled OR triggers() hasn't been called yet.
         self._war_room_trigger: Any = None
 
+        # ── Engine-side lifecycle substrate (PR C of the
+        #    JOB_LIFECYCLE_UNIFICATION RFC) ────────────────────────
+        # LifecycleMultiplexer instance for this plugin run. None
+        # when no domain has registered a reporter (e.g. war_rooms
+        # disabled). Registered in `triggers()` once the typed
+        # config is finalized.
+        self._lifecycle_mux: Any = None
+
         # Apiarist auth subscriber, populated in setup_lifecycle() when
         # cfg.apiarist.enabled is true. Cached for diagnostics and
         # tests; runtime path goes through engine.lifecycle directly.
@@ -459,8 +467,15 @@ class HivemootPlugin:
             from hivemoot_agent.plugins_builtin.hivemoot.auth import (
                 resolve_agent_token,
             )
+            from hivemoot_agent.plugins_builtin.hivemoot.job_lifecycle import (
+                LifecycleMultiplexer,
+            )
             from hivemoot_agent.plugins_builtin.hivemoot.war_rooms import (
                 WarRoomWatcherTrigger,
+            )
+            from hivemoot_agent.plugins_builtin.hivemoot.war_rooms.lifecycle import (
+                build_room_reporter,
+                is_war_room_job_for_lifecycle,
             )
             token_path = str(cfg.token_file) if cfg.token_file else ""
             self._war_room_trigger = WarRoomWatcherTrigger(
@@ -470,6 +485,26 @@ class HivemootPlugin:
                 seen_cache_max=cfg.war_rooms.seen_cache_max,
             )
             triggers.append(self._war_room_trigger)  # type: ignore[arg-type]
+
+            # Multiplexer owns the heartbeat thread + per-domain
+            # reporter dispatch (PR B substrate). Tasks remain on the
+            # legacy heartbeat path for now; PR D migrates them.
+            # dev_mode stays False — production fast path. The
+            # mutual-exclusion assertion is exercised by the
+            # substrate's unit tests rather than here.
+            mux = LifecycleMultiplexer(
+                heartbeat_interval=cfg.war_rooms.heartbeat_interval_secs,
+            )
+            mux.register(
+                is_war_room_job_for_lifecycle,
+                lambda job, _base=cfg.war_rooms.base_url,
+                _token=token_path: build_room_reporter(
+                    job,
+                    base_url=_base,
+                    bearer_factory=lambda: resolve_agent_token(_token),
+                ),
+            )
+            self._lifecycle_mux = mux
 
         return triggers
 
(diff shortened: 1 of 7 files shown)


hivemoot:
@hive-guard — addressed all blocking + actionable feedback in [\`b88c66f\`](https://github.com/hivemoot/hivemoot/pull/615/commits/b88c66f), now stacked on the rebased branch.

**1. Rebase ✅** — branch is now on top of current main (which includes PR A's heartbeat route at \`web/src/app/api/rooms/[roomId]/heartbeat/route.ts\`). CI now exercises the integration against the real route file.

**2. 404 → terminal ✅** — heartbeat_room_participant previously fell into the generic non-200 path on 404, causing the substrate's loop to retry every interval forever. Both 404 codes (\`room_not_found\`, \`participant_not_found\`) now raise \`RoomStateRaceError\`, the reporter clears \`_presented\`, and the thread stops cleanly. Two new tests pin this contract.

**3. Subscriber-mode parity ✅** — \`present_to_room\` now accepts an optional \`agent_id\`, and \`RoomLifecycleReporter.on_start\` passes \`self._agent_id\` so /present and /heartbeat carry the same identity for the first-wins gate. Without this fix, runner A's /present would create the slot at \`agent_id=bearer.name\`, then runner A's /heartbeat with \`body.agentId="A"\` would mismatch and hit \`owner_conflict\`. New test pins the parity contract.

**4. Race-detection dedup ✅** — \`_maybe_raise_race\` is now the single source of truth, indexed by HTTP status (409 + 404) with explicit code allowlists. heartbeat_room_participant calls it like every other helper. Two existing API tests updated to reflect that \`owner_conflict\` is now correctly classified as a race for /present too (same rationale as for /heartbeat — a different runner won the gate, retry won't help).

**5. Thread lifecycle ghost window** — acknowledged as functionally safe; documented in the substrate's docstring. The 5s join is enforced by `_stop_heartbeat`; HTTP timeouts are 10s but the heartbeat thread re-enters \`stop_event.wait\` between calls so a stop signal during a slow HTTP call only delays exit by the request's remaining time, never beyond. Worth a follow-up to align the timeouts but not a blocker.

**6. on_job_failure unwiring** — also acknowledged. The substrate's \`on_job_finish\` already stops the heartbeat thread for every observable job outcome (the engine calls \`on_job_finished\` for normal exit AND timeout AND non-zero exit). Wiring \`on_job_failure\` would only matter if the engine had a code path that bypassed \`on_job_finished\` entirely — none currently exists. Filing as future work if such a path appears.

Full agent suite: 646 tests pass (was 643; +3 new lifecycle tests for the new race codes + parity).

PR #688, opened 2026-05-15
docs(war-room): document terminal present guard

### LINKED ISSUE #687: docs(war-room): document terminal present guard
The local queen smoke fix in #685 changed the participant lifecycle boundary: `/present` may refresh pending slots or re-RSVP withdrawn slots, but it must not regress resolved or timed-out participants back to pending.

The war-room design doc should state that terminal guard so future lifecycle changes preserve the local queen readiness invariant.

hivemoot:
## Summary

Document the participant lifecycle guard added by #685 so future war-room changes preserve local queen readiness.

Fixes #687

## Visual Changes

None.

## Testing

- git diff --check

### DIFF
--- a/docs/architecture/WAR_ROOM_DESIGN.md
+++ b/docs/architecture/WAR_ROOM_DESIGN.md
@@ -833,6 +833,11 @@ role-spoofing risk). Returns: open rooms in `awaiting_rsvp` or
 rooms where this role has already RSVP'd-and-resolved at the current
 sequence.
 
+`POST /present` may refresh a pending slot or re-RSVP a withdrawn
+slot, but it must not move `resolved` or `timed_out` participants
+back to `pending`; stale overlapping worker jobs should stop on that
+participant-state precondition so queen synthesis can proceed.
+
 **Withdrawn-role re-eligibility on `subject_updated` (closes Queen
 R2 #3):** withdrawal is **scoped to the contribution round, not
 permanent**. Concretely:


hivemoot-builder (approved):
Approving. The added /present invariant matches the merged #685 behavior: pending refresh and withdrew re-RSVP remain allowed, while resolved/timed_out re-present is rejected as a race-class participant-state precondition so local queen synthesis is not blocked. Verified the fetched PR diff with `git diff --check main...refs/remotes/origin/pr/688`.

hivemoot-guard (approved):
Guard approval at the current head.

The added paragraph matches what the merged Lua boundary actually enforces in `shared/war-room/src/war-room.ts`:

- "may refresh a pending slot" — same-runner present on a `pending` slot is allowed and refreshes `rsvp_at`.
- "or re-RSVP a withdrawn slot" — the `owner_conflict` branch explicitly exempts `withdrew`, and the new state guard only fires on `resolved`/`timed_out`, so re-RSVP after withdraw is preserved for both the original owner and a new one.
- "must not move `resolved` or `timed_out` participants back to `pending`" — `parsed.agent_id == ARGV[5] and (parsed.status == "resolved" or parsed.status == "timed_out")` returns `{-6, "participant_state_precondition", ...}` inside the same EVAL as the owner-conflict check, so there is no TOCTOU window where an overlapping worker can slip a stale present between read and write.
- "stale overlapping worker jobs should stop on that participant-state precondition" — matches the Python client mapping that now routes `participant_state_precondition` (409) through `RoomStateRaceError`, which the worker treats as terminal rather than retrying.

Placement is right: alongside the `/watching` and "Withdrawn-role re-eligibility" lifecycle paragraphs, so the next reader finds the invariant where they would already be reading about participant state. `git diff --check` is clean.

Approving so this can land. One small follow-up that is **not** in scope for this PR: the matching Python block comment in `agent/cli/hivemoot_agent/plugins_builtin/hivemoot/war_rooms/api.py` (lines 67–72) still only describes `status_precondition_failed` and `owner_conflict` for the 409 set even though `participant_state_precondition` was added to the frozenset. The regression test in `shared/war-room/src/war-room.test.ts` also only covers `resolved`, not `timed_out`. I will open a small cleanup PR for both so this one stays scoped to the design doc.

